PR #3, opened 2025-01-20
Setup pinia

gaspergrom:
## Release Notes

- **New Features**
  - Added a new asynchronous action for token management within the module.
  - Introduced a new store module for organized state management.

- **Chores**
  - Updated project dependencies to include Pinia and ESLint.
  - Updated Nuxt configuration to support Pinia integration.

- **Configuration**
  - Added `.gitignore` updates for IDE-specific files.


frontend/components/modules/moduleName/store/moduleName.state.ts
```diff
@@ -0,0 +1,7 @@
+export interface ModuleNameState {
+  token: any,
```
emlimlf, line 2: Is it possible to add the type here instead of any?

PR #98, opened 2025-03-12
added all the metrics for the activity dropdown

emlimlf:
## In this PR

Added all the activity types on the Activity dropdown shared component. 

## Ticket
[INS-169](https://linear.app/lfx/issue/INS-169/show-all-activities-grouped-by-platform-on-the-activity-dropdown)

## Future Work
Filter the activities shown in this dropdown to just the activities that are enabled for the specific project

## Summary by CodeRabbit

- **New Features**
	- Enhanced the metrics selector with grouped categories for a clearer, more organized view of available options.
	- Updated the display behavior of dropdown options for improved visual clarity and user experience.
	- Changed the default active tab for displaying active contributors and organizations, potentially altering the initial view presented to users.

- **Bug Fixes**
	- Improved observability for debugging purposes by adding console logging to monitor changes in the active tab.

### DIFF
--- a/frontend/app/layouts/default.vue
+++ b/frontend/app/layouts/default.vue
@@ -3,35 +3,6 @@
     <lfx-navbar />
     <slot />
     <lfx-toast theme="dark" />
-    <div class="px-10 mx-auto max-w-3xl pb-6 pt-6">
-      <div class="copyright-container">
-        <p
-          class="copyright text-xs text-center"
-        >
-          Copyright © 2025 The Linux Foundation®. All rights reserved.
-          The Linux Foundation has registered trademarks and uses trademarks. For more details, including terms of use,
-          <a
-            href="https://www.linuxfoundation.org/legal/platform-use-agreement/"
-            target="_blank"
-            rel="noopener noreferrer"
-          >Platform Usage</a>,
-          <a
-            href="https://www.linuxfoundation.org/legal/privacy-policy?hsLang=en"
-            target="_blank"
-            rel="noopener noreferrer"
-          >Privacy Policy</a>,
-          and <a
-            href="https://www.linuxfoundation.org/legal/trademark-usage?hsLang=en"
-            target="_blank"
-            rel="noopener noreferrer"
-          >Trademark Usage</a>,
-          please see our <a
-            href="https://www.linuxfoundation.org/legal/policies"
-            target="_blank"
-            rel="noopener noreferrer"
-          >Policies</a> page.</p>
-      </div>
-    </div>
   </main>
 <!--  <lfx-footer />-->
 </template>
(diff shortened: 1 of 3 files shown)


frontend/app/components/modules/project/components/contributors/active-contributors.vue
```diff
@@ -121,6 +121,7 @@ const barChartConfig = computed(() => getBarChartConfig(chartData.value, chartSe
 
 watch(selectedTimeRangeKey, () => {
   activeTab.value = tabs.value[0]?.value || 'weekly';
+  console.log('activeTab', activeTab.value);
```
borfast: Looks like this one got left behind.

frontend/app/components/modules/project/components/contributors/config/metrics.ts
```diff
@@ -1,13 +1,378 @@
...
-  { label: 'All activities', value: 'all' },
-  { label: 'Commits', value: 'commits' },
-  { label: 'Issues Opened', value: 'issues-opened' },
-  { label: 'Issues Closed', value: 'issues-closed' },
-  { label: 'Pull Requests Opened', value: 'pull-requests-opened' },
-  { label: 'Pull Requests Closed', value: 'pull-requests-closed' },
-  { label: 'Pull Requests Merged', value: 'pull-requests-merged' },
-  { label: 'Pull Requests Reviews', value: 'pull-requests-reviews' },
-  { label: 'Pull Requests Comments', value: 'pull-requests-comments' }
+export const metricsOptions: DropdownGroupOptions[] = [
```
borfast, line 3: Do we have a bit of a mismatch in terms of language here? We talk about "activities" but we're calling them "metrics" in the code, which can be a little confusing. A metric would be *the number* of a thing, not the thing itself.  And we are filtering by the thing, not its number. Does that make sense, or am I mixing things?

frontend/app/components/modules/project/components/contributors/config/metrics.ts
```diff
@@ -1,13 +1,378 @@
-import type { DropdownOption } from '~/components/uikit/dropdown/types/dropdown.types';
+import type { DropdownGroupOptions } from '~/components/uikit/dropdown/types/dropdown.types';
```
borfast, line 1: This would be for a separate PR but, since the type `DropdownGroupOptions` (and the related `DropdownOption`)  is specifically about activities and not just a generic dropdown widget, wouldn't it be better to name it something like `ActivitiesDropdownGroupOptions`, or something like that more explicit and less generic?

emlimlf: No, DropdownGroupOptions is generic to the shared select component.

PR #155, opened 2025-04-07
fix the default start and end date on project store

emlimlf:
## In this PR

Fix the default start and end date that was set on the project store. The issue wasn't with granularity but with this default start and end date that the widgets are picking up.

## Ticket
[INS-384](https://linear.app/lfx/issue/INS-384/default-granularity-not-being-applied)

## Summary by CodeRabbit

- **Refactor**
	- Updated how default time range and date options are initialized, ensuring that the displayed start and end dates are more accurately set. This improvement refines the filtering experience for selecting date ranges.
  
- **Style**
	- Adjusted the formatting of displayed project details for enhanced clarity.

### DIFF
--- a/frontend/app/components/modules/project/components/development/average-time-to-merge.vue
+++ b/frontend/app/components/modules/project/components/development/average-time-to-merge.vue
@@ -101,9 +101,9 @@ const averageTimeMerge = computed<AverageTimeMerge>(() => data.value as AverageT
 const summary = computed<Summary>(() => averageTimeMerge.value.summary);
 const chartData = computed<ChartData[]>(
   // convert the data to chart data
-  () => convertToChartData((averageTimeMerge.value?.data || []) as RawChartData[], 'dateFrom', [
+  () => convertToChartData((averageTimeMerge.value?.data || []) as RawChartData[], 'startDate', [
     'averageTime'
-  ], undefined, 'dateTo')
+  ], undefined, 'endDate')
 );
 
 const chartSeries = ref<ChartSeries[]>([
--- a/frontend/app/components/modules/project/components/development/issues-resolution.vue
+++ b/frontend/app/components/modules/project/components/development/issues-resolution.vue
@@ -47,7 +47,7 @@
               height="1.25rem"
               width="4rem"
             >
-              <span class="text-xl">{{ formatNumber(summary.avgVelocityInDays) }} days</span>
+              <span class="text-xl">{{ avgVelocityInDays }} days</span>
             </lfx-skeleton-state>
           </div>
         </div>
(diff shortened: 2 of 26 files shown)


frontend/app/components/modules/project/store/project.store.ts
```diff
@@ -37,11 +37,22 @@ const calculateGranularity = (start: string | null, end: string | null): string[
 export const useProjectStore = defineStore('project', () => {
   const route = useRoute();
 
-  const selectedTimeRangeKey = ref<string>(dateOptKeys.past365days);
-  const startDate = ref<string | null>(lfxProjectDateOptions[0]?.startDate || null);
-  const endDate = ref<string | null>(lfxProjectDateOptions[0]?.endDate || null);
+  const defaultTimeRangeKey = dateOptKeys.past365days;
+  const defaultDateOption = lfxProjectDateOptions.find(
+    (option) => option.key === defaultTimeRangeKey
```
borfast, line 42: Maybe I'm missing some Vue-specific knowledge here but this seems rather convoluted for something seemingly so simple :thinking:  Couldn't we simply use an Enum for the possible time ranges?

frontend/app/components/modules/project/store/project.store.ts
```diff
@@ -37,11 +37,22 @@ const calculateGranularity = (start: string | null, end: string | null): string[
...
+  );
+
+  const selectedTimeRangeKey = ref<string>(defaultTimeRangeKey);
+  const startDate = ref<string | null>(
+    defaultDateOption?.startDate || lfxProjectDateOptions[1]?.startDate || null
```
borfast, line 47: I just realised that the startDate and endDate can be null, even before these changes, but this shouldn't happen, the API should always receive both values. These and the project are the three parameters that the API always needs.

frontend/app/components/modules/project/store/project.store.ts
emlimlf: The dropdown options are in array. I don't want to touch that part as that is used by the date time range control in the header. This was originally written by @gaspergrom, I also think it's not a good idea to use enums there. Additionally, if we are to change this to enum, we also have to touch that part of the header date range control.

emlimlf: Besides, we do use enums for the timeRangeKeys

frontend/app/components/modules/project/store/project.store.ts
emlimlf: I thought the "All Time" key range don't need to pass a date hence the null values. I will have to loop in @gaspergrom for this, as he was the one who wrote this originally.

borfast: Then let's not block the issue for this. Let's capture it in a ticket in Linear so we don't forget it, and move on. I'll take care of it.

Edit: ticket [here](https://linear.app/lfx/issue/INS-404/ensure-both-start-and-end-dates-are-always-sent-from-the-frontend).

PR #318, opened 2025-06-03
added delay to set initial value for active item on the widget area

emlimlf:
## In this PR

Re-added the mechanism that detects the active widget from the URL params that got lost during the transition to widgets area components.

## Ticket

[INS-615](https://linear.app/lfx/issue/INS-615/insights-page-anchors-not-working)

- `main` 
  - \#318 :point\_left:
    - \#320

### DIFF
--- a/frontend/app/components/modules/project/components/shared/header.vue
+++ b/frontend/app/components/modules/project/components/shared/header.vue
@@ -108,7 +108,9 @@ SPDX-License-Identifier: MIT
           :class="scrollTop > 50 ? 'md:py-4' : 'md:py-5'"
         >
           <lfx-project-menu :project="props.project" />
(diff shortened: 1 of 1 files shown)


frontend/app/components/modules/widget/components/shared/widget-area.vue
```diff
@@ -113,6 +113,14 @@ const onBenchmarkUpdate = (value: Benchmark | undefined) => {
     benchmarks.value[value.key] = value;
   }
 }
+
+onMounted(() => {
+  // add a pause to wait for the page to load then set the active item
+  setTimeout(() => {
```
gaspergrom: Can you try with await nextTick() if it works. this waits for render. Otherwise i would have cleaner solution like watching for having correct project loaded and then navigate

PR #319, opened 2025-06-03
change the logic used to display the chart in the retention widget

emlimlf:
## In this PR

- Changed the widget name to Quarterly Retention
- Changed the granularity used by the chart from dynamic to fixed Quarterly
- Changed empty state check: the chart will now only appear if the selected time range is greater than 180days (2 quarters)

## Ticket
[INS-603](https://linear.app/lfx/issue/INS-603/display-an-empty-state-in-retention-widget-below-a-specific-period)


frontend/app/components/modules/widget/config/contributor/retention/retention.vue
```diff
@@ -94,45 +94,47 @@ const {
...
-      return [Granularity.YEARLY];
-    default:
-      return [Granularity.YEARLY];
-  }
-};
-const isSingleDay = computed(() => {
+// We might revisit this in the future
+// const calculateGranularity = (start: string | null, end: string | null): string[] => {
+//   // Return weekly if either date is null
+//   if (!start || !end) {
```
gaspergrom: If we wont need this block of code in the future please remove it

emlimlf: done

PR #320, opened 2025-06-03
Improvement/time range url param

emlimlf:
## In this PR

- Save the time range and dates when the user changes the selection
- Restore the time range and dates if it's available from the URL Params

## Ticket

- `main` 
  - \#318
    - \#320 :point\_left:

### DIFF
--- a/frontend/app/components/modules/project/store/project.store.ts
+++ b/frontend/app/components/modules/project/store/project.store.ts
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: MIT
 import { defineStore } from 'pinia';
 import { ref, computed } from 'vue';
-import { useRoute, useRouter } from 'nuxt/app';
+import { useRoute } from 'nuxt/app';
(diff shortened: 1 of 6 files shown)


frontend/app/components/modules/project/store/project.store.ts
```diff
@@ -40,21 +40,54 @@ export const defaultTimeRangeKey = dateOptKeys.past365days;
...
+
+export const updateUrlParams = (timeRange: string, start: string | null, end: string | null) => {
+  const route = useRoute();
+  const router = useRouter();
+
+  const query: Record<string, string | null | undefined> = {
+    ...route.query,
+    timeRange,
+    start: start || undefined,
+    end: end || undefined,
```
gaspergrom: This is good. i think would be good to make a reusable function just so we can use it on easily on other places aswell. For example if you will need on oss index to have filters in url query

emlimlf: done

PR #325, opened 2025-06-03
feat: insights temporal workers

epipav:
- `main` 
  - \#325 :point\_left:
    - \#405

### DIFF
--- a/frontend/app/assets/styles/base.scss
+++ b/frontend/app/assets/styles/base.scss
@@ -9,6 +9,8 @@ html {
   scroll-behavior: smooth;
   -webkit-overflow-scrolling: touch;
   position: relative;
+  overflow-x: hidden;
+  overflow-y: scroll;
 }
 
 body {
--- a/frontend/app/components/modules/explore/components/explore-header.vue
+++ b/frontend/app/components/modules/explore/components/explore-header.vue
@@ -15,7 +15,7 @@ SPDX-License-Identifier: MIT
 
       <div class="w-full flex justify-center mt-5 pb-10">
         <div
-          class="w-full sm:max-w-120 overflow-hidden bg-white h-[45px] border
+          class="w-full sm:max-w-120 overflow-hidden bg-neutral-50 h-[45px] border
           border-neutral-200 rounded-full flex items-center gap-2 px-5 cursor-pointer"
           @click="isModalOpen = true"
         >
--- a/frontend/app/components/modules/explore/components/list-card.vue
+++ b/frontend/app/components/modules/explore/components/list-card.vue
@@ -34,6 +34,7 @@ SPDX-License-Identifier: MIT
   <lfx-explore-list-modal
     v-model="isModalOpen"
     :value="props.value"
+    width="27rem"
   />
 </template>
 
--- a/frontend/app/components/modules/explore/components/source-index-progress.vue
+++ b/frontend/app/components/modules/explore/components/source-index-progress.vue
@@ -31,7 +31,7 @@ SPDX-License-Identifier: MIT
         </div>
       </div>
     </div>
-    <div class="flex justify-center sm:justify-start">
+    <div class="flex justify-center sm:justify-start py-2">
       <nuxt-link
         :to="exploreLink"
         class="text-sm text-brand-500 font-semibold hover:underline"
--- a/frontend/app/components/modules/explore/components/top-section.vue
+++ b/frontend/app/components/modules/explore/components/top-section.vue
@@ -4,7 +4,7 @@ SPDX-License-Identifier: MIT
 -->
 <template>
   <section>
-    <div class="lg:flex hidden flex-row gap-8 items-start">
+    <div class="lg:flex hidden flex-row gap-4 xl:gap-8 items-start">
       <lfx-explore-list-card
         v-for="tab in TOP_SECTION_TABS"
         :key="tab.title"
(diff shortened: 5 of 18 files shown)


scripts/services/insights-app.yaml
```diff
@@ -6,65 +6,26 @@ x-env-args: &env-args
...
-insights-app:
-  api:
+services:
+  insights-app:
     build:
       context: ../../
-      dockerfile: ./scripts/services/docker/Dockerfile.backend
-    command: 'pnpm run start:api'
-    working_dir: /usr/crowd/app/backend
+      dockerfile: ./scripts/services/docker/Dockerfile.insights_app
```
borfast: nit: I would move the Docker stuff to its own folder, since it's not really a service in the sense of something that we want running on our infrastructure, serving requests, doing scheduled tasks, etc.

And I would suggest moving it to a `docker` directory inside the `frontend` directory, and name it just `Dockerfile`. That way things are isolated within their own sphere of action. Kind of like the locality of behaviour principle for code.

.env.dist.composed
```diff
@@ -0,0 +1,3 @@
+INSIGHTS_DB_READ_HOST=db
```
borfast: I always thought this pattern we have on crowd.dev of the .dist.local and .dist.composed files is super confusing and I don't think I ever understood why it is necessary. Could we perhaps simplify this to the more common pattern of having a single .env.dist file with some default values and add instructions in the README for the user to copy the file to .env and change values as necessary?

If we really want to stick to this pattern, then we should at least add a clear explanation of how it works.

scripts/cli
```diff
@@ -0,0 +1,698 @@
+#!/usr/bin/env bash
```
borfast: This is one I would really, really love we didn't use. I didn't read the whole script, just skimmed the first few dozen lines, but it seems to be pretty much the same thing as the one we have in crowd.dev, so my comment is mostly based on my experience with that one.

The problem is that the script hides away too much, makes it super hard to understand what is being done when running a command (we have to follow multiple function calls, sometimes to different files), and if we're using Docker Compose, most of this isn't necessary anyway.

Instead of having to learn how to use a custom script, we should know how to use the foundational tools (i.e. Docker, Docker Compose, etc).

In fact, some of this might even cause issues, like explicitly specifying an IP range for the docker network. Docker does this automatically for us and doesn't clash with existing networks, and doesn't make our lives harder if we want to do the networking slightly different for some reason, like debugging something.

If we really must have scripts that hide away what we're doing, can we at least separate them into individual files for each functionality, so that the code is at least a little bit easier to digest?

scripts/scaffold.yaml
```diff
@@ -0,0 +1,32 @@
+services:
```
borfast: Same suggestion as before, about moving this to `frontend/docker`, and also I would suggest this should be given the standard name `docker-compose.yaml` to avoid confusion.

workers/base/base.tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
```
borfast: We may want to have other "workers", so perhaps we could have this inside a directory for temporal, maybe inside the services folder?

borfast:
In general it looks ok, but I think we could improve the code organisation. I know you started by trying to mimic the structure we have on crowd.dev, but what we have there is a bit confusing and "entangled" with all the indirections between docker files, for example.

So my suggestions would be to:

1. Not use the `.env.dist.composed` pattern and instead use the more common pattern of having a single `.env.dist` file for each service with default values that can be public and removing those that can´t. The user should copy the file to `.env` and replace the necessary values.
2. Not use the CLI as it currently stands, because it hides away too much, makes understanding what happens harder than it needs to be, and running the underlying tools isn't that hard anyway. If we really want to have scripts abstract away the underlying tools, then we should make sure there is good documentation about the tool and perhaps also split it into multiple scripts with explicit names for their functionality.
3. Use standard names for docker-related files, like `Dockerfile` and `docker-compose.yaml`, so that it is immediately clear what they are, and move them into a `docker` folder inside each service directory, instead of having them all together inside a single folder.
4. Move the `workers` directory into a Temporal-dedicated folder, like `services/temportal/workers`.

database/flyway_migrate.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
```
borfast: I forgot this one: if there's no particular reason to have a Dockerfile here, which means we have to build a custom image just to run migrations, we could run a container directly from the official image.

The bash script could become something like this:
```bash
#!/usr/bin/env bash
set -ex
set +o history

echo "Migrating jdbc:postgresql://${PGHOST}:${PGPORT}/${PGDATABASE}"

docker run --rm \
  -v "$(pwd)/database/migrations:/tmp/migrations" \
  flyway/flyway:7.8-alpine \
  -locations="filesystem:/tmp/migrations" \
  -url="jdbc:postgresql://${PGHOST}:${PGPORT}/${PGDATABASE}" \
  -user="$PGUSER" \
  -password="$PGPASSWORD" \
  -connectRetries=60 \
  -outOfOrder=true \
  -placeholderReplacement=false \
  -schemas=public \
  -X \
  migrate

set -o history
```

One less file to manage, and one less container image cluttering our hard drives.

workers/base/base.tsconfig.json
epipav: This is a reusable tsconfig for all temporal workers. The package downloads worker currently uses this file for its tsconfig

database/flyway_migrate.sh
epipav: added migrate.sh here while removing the dockerfile, and also a create_migration.sh for eash file generation :+1:

PR #327, opened 2025-06-04
reset query params when changing projects

emlimlf:
## In this PR

- Reset the query params when the user changes projects. It was causing an incorrect time range value if it's not reset
- Fix typo in the benchmark text on overview page

## Ticket


frontend/app/components/modules/project/components/shared/header/date-range-picker.vue
```diff
@@ -133,7 +133,7 @@ const { selectedTimeRangeKey, startDate, endDate } = storeToRefs(useProjectStore
 const { queryParams } = useQueryParam();
 const isOpen = ref(false);
 const isCustomSelectorOpen = ref(false);
-
+console.log("!!!selected", selectedTimeRangeKey.value);
```
gaspergrom: Remove console log

PR #355, opened 2025-06-10
fix custom date url param selection

emlimlf:
## In this PR 

Added check for custom date time range in URL param and use the date in the query instead of default

## Ticket

### DIFF
--- a/frontend/app/components/shared/utils/query-param.ts
+++ b/frontend/app/components/shared/utils/query-param.ts
@@ -69,18 +69,24 @@ export const useQueryParam = () => {
         // get the start and end date option from the time range
         const { start: startOption, end: endOption } = getStartAndEndDate(timeRange);
 
-        start = isValidStartDate
-          ? startOption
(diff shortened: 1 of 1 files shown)


frontend/app/components/shared/utils/query-param.ts
```diff
@@ -69,18 +69,24 @@ export const useQueryParam = () => {
...
-
-        end = isValidEndDate
-          ? endOption
-          : defaultDateOption?.endDate || lfxProjectDateOptions[1]?.endDate || null;
+        if (timeRange === dateOptKeys.custom) {
+          start = paramStart as string;
+          end = paramEnd as string;
+        } else {
+          start = isValidStartDate
+            ? startOption
```
gaspergrom: Since its a common function i wouldnt stictly check certain parameters here, i would make a callback function to check for valid query params and pass it to the common function

gaspergrom (changes requested):

PR #368, opened 2025-06-11
feat: decouple tinybird.ts from Nuxt [INS-659]

borfast:
In order to use the code that fetches data from Tinybird in other parts of the project other than the backend, we need to decouple the tinybird code from Nuxt, namely by not using `useRuntimeConfig`.

- `main` 
  - \#368 :point\_left:
    - \#369
      - \#371


emlimlf (approved):

gaspergrom (approved):
looks good!

PR #369, opened 2025-06-11
feat: fetch search data from keywords everywhere [INS-659]

borfast:
[Ticket in Linear](https://linear.app/lfx/issue/INS-659/implement-the-code-that-will-fetch-the-data-from-keywords-everywhere)

This adds the initial implementation of the code that fetches search volume data from Keywords Everywhere.

- `main` 
  - \#369 :point\_left:
    - \#371


borfast:
I'm going to pause this one because we have other big changes coming in that will affect the project structure, and I'll redo this later to match those changes.

PR #398, opened 2025-06-18
fix: sharing modal opening after native one

gaspergrom:


gaspergrom:
https://linear.app/lfx/issue/INS-702/share-modal-opens-after-native-sharing

PR #403, opened 2025-06-18
fix: issues resolution header broken on mobile

gaspergrom:


gaspergrom:
https://linear.app/lfx/issue/INS-697/issues-resolution-header-is-broken-on-mobile

PR #409, opened 2025-06-18
fix: explore page fixes

gaspergrom:


gaspergrom:
This fixes
https://linear.app/lfx/issue/INS-693/tabs-are-cut-out-on-mobile

and other issues i noticed

PR #417, opened 2025-06-19
fix: reset selected time filters when the user changes project

emlimlf:
## In this PR

Set the time range and dates to default whenever the user changes project selection

## Ticket
[INS-624](https://linear.app/lfx/issue/INS-624/filters-should-not-persists-across-projects)


gaspergrom (approved):

gaspergrom:
looks good!

PR #420, opened 2025-06-19
fix: explore projects contributor count

gaspergrom:


gaspergrom:
Issue: https://linear.app/lfx/issue/INS-710/use-contributor-count-instead-of-criticality-score-for-leaderboard

emlimlf (approved):

PR #422, opened 2025-06-19
fix: security overview tab layout

gaspergrom:


gaspergrom:
Issue: https://linear.app/lfx/issue/INS-695/security-tab-broken-on-mobile

PR #424, opened 2025-06-19
fix: project header breaking

gaspergrom:


gaspergrom:
Issues:
https://linear.app/lfx/issue/INS-707/all-repositories-goes-into-next-line
https://linear.app/lfx/issue/INS-704/selected-repository-is-cutout

PR #430, opened 2025-06-19
fix: mobile widget menu on tablet

gaspergrom:


gaspergrom:
This fixes issue that menu can be opened only on hover on tablet

PR #446, opened 2025-06-23
feat: cache pages and api calls only on production

gaspergrom:


gaspergrom:
https://linear.app/lfx/issue/INS-715/remove-caching-on-staging-and-remove-redis

PR #451, opened 2025-06-23
chore: use pnpm in isolated mode for frontend

epipav:
removes ./frontend from pnpm-workspace.yaml - instead we'll use it in isolated mode (by it's own pnpm-lock.yaml) until when/if we need shared libraries from the nuxt app

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -18,190 +18,6 @@ importers:
         specifier: ^15.5.2
         version: 15.5.2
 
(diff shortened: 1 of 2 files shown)


gaspergrom (approved):

borfast:
We already need this, though it's for a small thing. We use the TinyBird library for the search volume worker. Currently, I copied the library into the service worker directory, but this is meant to be a temporary solution.

PR #470, opened 2025-06-27
fix: url parameter history

emlimlf:
## In this PR

- Changed the way the URL params are stored by using router.push instead of replace, this enables the browser to store the filter changes and be able to use the back or forward button
- Applied the changes to the projects page and on the collections list page

## Ticket
[INS-631](https://linear.app/lfx/issue/INS-631/url-history-doesnt-seem-to-be-used-when-changing-the-date-range)


gaspergrom (approved):
Looks good

PR #476, opened 2025-07-01
feat: hide empty columns for package downloads and dependency charts

emlimlf:
## In this PR

- Hide empty columns for both the package downloads and dependency widgets
- Added "We don't have data before June 2025." to the description for both widgets
- Fix summary display for the dependency chart that overlaps for large values

## Ticket
[INS-732](https://linear.app/lfx/issue/INS-732/exclude-all-periods-before-june-2025-in-package-downloadsdependencies)


gaspergrom (approved):
Looks good

PR #490, opened 2025-07-08
fix: undefined title in reported jira issues

emlimlf:
## In this PR

- Created a config file that contains page title and description for explore, collections and open source index pages to be used in the respective pages.
- Added the page title as part of the data sent when reporting an issue
- Used the page title as a fall back when the project name isn't set on the create jira issue endpoint

## Ticket 
[INS-753](https://linear.app/lfx/issue/INS-753/issue-reported-from-footer-on-oss-index-ads-undefined-title)


frontend/app/config/pageTitles.ts
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) 2025 The Linux Foundation and each contributor.
+// SPDX-License-Identifier: MIT
+
+export const pageTitlesDescriptions = {
```
gaspergrom: This is not well it missess all dynamic pages like collection details and subpages of the oss index so make it dynamic by accessing document.title

emlimlf: The collections details already has a dynamic page title setting, see pages/collection/[slug].vue line 54 onwards. As for the OSS Index, I pushed a fix that address that.

gaspergrom: dont use configuration like this to have page titles done. just inject document.title into data sent and use that

gaspergrom: if we add new page you will have unnecessary aditional work

emlimlf: done, removed the page title config

gaspergrom (approved):

PR #521, opened 2025-07-17
docs: maintainers

joanagmaia:


joanagmaia:
Hey @jonathimer, can you check this one out? We have received a couple of questions now around this, and how users want to update the data. We should be transparent about this as soon as possible.

PR #565, opened 2025-08-01
chore: replaced lodash plugin with the es version

emlimlf:
## In this PR

- Replaced the lodash plugin with lodash-es
- Updated all the usage to use the tree-shaken method

## Ticket
[INS-775](https://linear.app/lfx/issue/INS-775/chore-update-lodash-plugin)


gaspergrom (approved):
Thanks for taking care of this it was needed

PR #575, opened 2025-08-06
feat: insights chat

epipav:


epipav:
#600 is the PR with squashed/signed-off commit since the PR had few commits without sign-off

PR #600, opened 2025-08-26
feat: data copilot beta

epipav:


frontend/app/components/shared/utils/jwt-permissions.ts
```diff
@@ -0,0 +1,53 @@
...
+
+const isLocal = (): boolean => {
+  const config = useRuntimeConfig()
+  const appEnv = config.public.appEnv
+
+  return appEnv !== 'staging' && appEnv !== 'production'
+}
+
+const GROUP_CLAIM_KEY = 'https://sso.linuxfoundation.org/claims/groups'
+const GROUP_NAME = 'lfproducts-lfx-insights'
```
joanagmaia, line 14: Should we include this in the env file?

epipav: makes sense :+1:

PR #620, opened 2025-09-02
fix: fix "includes archived repos" tag being shown when no archived repos are selected

borfast:
When selecting repositories in the search list and no archived repository is selected, the "includes archived repositories" tag is still being shown.

image

The condition to show the tag was checking just if the project itself contained archived repositories, not if there were archived repositories in the list of selected ones.

This also affects other places, like the security tab and the project overview, where the indication of archived repositories is also shown.


joanagmaia (approved):
LGTM 🚀 
Thank you for helping with the fix 🙏

PR #642, opened 2025-09-05
docs: update docs to reflect excluded repositories in health and security metrics

borfast:
Now that we are excluding archived repositories from health and security scores, we want to reflect this in the project documentation.

This PR updates the two relevant pages.

[Ticket in Jira](https://linuxfoundation.atlassian.net/browse/IN-576)


joanagmaia (commented):
If we can add just a small mention to the "excluded" repos as well it would be great. As for the archived ones the added docs look good 👌

PR #709, opened 2025-10-05
Chore/use specific pipe for top projects

borfast:
This uses the new pipe created [here](https://github.com/CrowdDotDev/crowd.dev/pull/3451) for the Top LF projects, so that we can filter the data by activity types.


gaspergrom (approved):

borfast:
We decided to do this differently and filter on services/libs/tinybird/pipes/segmentId_aggregates_mv.pipe, which has the advantage that it covers other downstream pipes as well. The change was done [here](https://github.com/CrowdDotDev/crowd.dev/pull/3432/commits/10ed24a8b7f2dfbb36dd282664df24ca75050435).
Closing this one.

PR #716, opened 2025-10-07
feat: security yaml generation

gaspergrom:


frontend/app/components/modules/project/components/security/yaml/generate-yaml-modal.vue
```diff
@@ -0,0 +1,236 @@
...
+    const link = document.createElement('a')
+    link.href = url
+    link.download = 'security.yaml'
+    document.body.appendChild(link)
+    link.click()
+    document.body.removeChild(link)
+    URL.revokeObjectURL(url)
+    showToast('YAML file successfully downloaded', ToastTypesEnum.positive, 'circle-check')
+  } catch (error) {
+    console.error('Failed to download YAML file:', error)
```
emlimlf, line 205: I think you should use toast here to let the user know that the download YAML file failed

emlimlf: You should also URL.revokeObjectURL(url) when it fails

PR #724, opened 2025-10-08
feat: add dynamic activity type filtering when toggling collaborations

emlimlf:
## In this PR

- Added new flag to the activity types in the platform configs for the collaboration flag
- Used new flag in the activity types dropdown to filter out collaborations only types if the the include collaborations toggle is set to false
- Reset the selected activity type if the current one is not included in the activity types list when the user toggles the include collaborations back to false

## Ticket
[N-760](https://linuxfoundation.atlassian.net/browse/IN-760)

### DIFF
--- a/frontend/app/config/platforms/configs/github.platform.ts
+++ b/frontend/app/config/platforms/configs/github.platform.ts
@@ -1,8 +1,8 @@
 // Copyright (c) 2025 The Linux Foundation and each contributor.
 // SPDX-License-Identifier: MIT
-import { ActivityPlatforms } from '~~/types/shared/activity-platforms';
-import { ActivityTypes } from '~~/types/shared/activity-types';
-import type { PlatformConfig } from '~~/types/shared/platforms.types';
+import { ActivityPlatforms } from '~~/types/shared/activity-platforms'
+import { ActivityTypes } from '~~/types/shared/activity-types'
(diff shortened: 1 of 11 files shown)


frontend/app/config/platforms/configs/github.platform.ts
```diff
@@ -11,52 +11,57 @@ export const github: PlatformConfig = {
...
+      label: 'Started a discussion',
+      isCollaborationType: true,
     },
     {
       key: ActivityTypes.DISCUSSION_COMMENT,
-      label: 'Commented on a discussion'
-    }
-  ]
-};
+      label: 'Commented on a discussion',
```
gaspergrom, line 63: Since we have this in database now we could also add label to database to simply control it on the frontent maybe ask joana if this makes sense. Is contribution i think its also in activity types database

emlimlf: @joanagmaia ^

joanagmaia: Yes @gaspergrom, Raúl is gonna handle that new endpoint next. It's something that he had already started working on that's why I didn't ask Efren.

This PR/change is a temporary one until we have the endpoint because right now the activity types are wrong.

gaspergrom: sounds good

PR #733, opened 2025-10-10
feat: hide activity types that don't exist in project [IN-751]

borfast:
In the widgets that currently support filtering by activity type (see image below for an example), we want to change the available options, so that activity types that do not exist for the displayed project are not shown as filter options.

image

This is hard-coded in the platform config files [here](https://github.com/linuxfoundation/insights/tree/58d116bac9b20cb77a14d4fb6b75377aeb073aa7/frontend/app/config/platforms/configs).

We want to change this so that only activity types that exist for the given project are shown - i.e. if a project has no github issues, this activity type should not be displayed in the dropdown as a filtering option.

In this PR we:
* create an API endpoint for the frontend to get this information;
* the corresponding Tinybird data source that fetches the required data;
* change the frontend so that it fetches the list of activity types to display from Tinybird (via the backend API), which filters them at the source based on the provided project and, optionally, repositories.

Since the activity types from the platforms config files are not necessary or even wanted anymore, they are also being removed.

This depends on my two other related PRs from crowd.dev to be merged and deployed first ([3488](https://github.com/CrowdDotDev/crowd.dev/pull/3488) and [3489](https://github.com/CrowdDotDev/crowd.dev/pull/3489)).

Jira tickets:
* [IN-511](https://linuxfoundation.atlassian.net/browse/IN-511) (for adding the backend endpoint)
* [IN-751](https://linuxfoundation.atlassian.net/browse/IN-751) (for modifying the frontend)

These were initially two separate PRs but to avoid some issues and make it easier to review (despite this way having more files in a single PR) I merged them into a single one, and so I'm closing #745 in favour of this one.

- `main` 
  - \#745
    - \#733 :point\_left:

[IN-511]: https://linuxfoundation.atlassian.net/browse/IN-511?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #745, opened 2025-10-14
feat: provide a new endpoint with available activity types [IN-511]

borfast:
In the widgets that currently support filtering by activity type (see image below for an example), we want to change the available options, so that activity types that do not exist for the displayed project are not shown as filter options.

image

This PR creates an API endpoint for the frontend to get this information, and the corresponding Tinybird data source that fetches the required data.

[Jira ticket](https://linuxfoundation.atlassian.net/browse/IN-511)

- `main` 
  - \#745 :point\_left:
    - \#733


borfast:
This is related to #733, as they were initially intended to be two separate PRs, but to avoid some issues and make it easier to review (despite this way having more files in a single PR) I merged them into a single one, and so I'm closing this one in favour of the former.

PR #1342, opened 2025-10-31
Change color of unmeasurable security items from warning to neutral

### LINKED ISSUE #618: Change color of things you can't measure to grey / white
If someone looking at LFX insights sees orange/yellow when looking at an item, the implication is there is something wrong.  A lot of the stuff related to baseline you frequently cannot accurately measure automatically.  I would suggest you change the color here to grey or white if you cannot measure it, but you reasonably expect it may not be a problem.

I would also suggest that the score not be lowered for these items.  It should be treated as "cannot accurately measure" instead of "is missing".

balaraj74:
## Summary

This pull request addresses issue #618 by changing the visual representation of security items that cannot be accurately measured. Previously, these items were displayed with an orange/yellow warning badge, which gave users the impression that something was wrong. The fix changes the color to a neutral grey, making it clear that these items simply cannot be measured rather than indicating a problem.

## Changes Made

- Updated the NEEDS_REVIEW status tag to use neutral variation instead of warning-solid
- Changed the icon from exclamation-triangle to question-circle for better clarity
- Updated the display text from "Needs review" to "Cannot measure" to be more explicit
- Verified that score calculation already excludes these items from affecting the overall score

## Testing

- Tested the component in isolation to verify the visual changes
- Confirmed that the grey color no longer implies a warning or error state
- Verified that items marked as "Cannot measure" are excluded from score calculations

## Screenshots

The fix changes the appearance from an orange warning badge to a grey neutral badge, removing any negative implication for items that cannot be measured.

Fixes #618

### DIFF
--- a/frontend/app/components/modules/project/components/security/evaluation-result-tag.vue
+++ b/frontend/app/components/modules/project/components/security/evaluation-result-tag.vue
@@ -49,7 +49,7 @@ const typeData: Record<SecurityDataResult, TypeData> = {
     text: 'Cannot measure',
   },
 };
-const data = computed<number>(() => typeData[props.result]);
+const data = computed<TypeData>(() => typeData[props.result]);
 </script>
 
 <script lang="ts">


balaraj74:
> ## Summary
> This pull request addresses issue #618 by changing the visual representation of security items that cannot be accurately measured. Previously, these items were displayed with an orange/yellow warning badge, which gave users the impression that something was wrong. The fix changes the color to a neutral grey, making it clear that these items simply cannot be measured rather than indicating a problem.
> 
> ## Changes Made
> * Updated the NEEDS_REVIEW status tag to use neutral variation instead of warning-solid
> * Changed the icon from exclamation-triangle to question-circle for better clarity
> * Updated the display text from "Needs review" to "Cannot measure" to be more explicit
> * Verified that score calculation already excludes these items from affecting the overall score
> 
> ## Testing
> * Tested the component in isolation to verify the visual changes
> * Confirmed that the grey color no longer implies a warning or error state
> * Verified that items marked as "Cannot measure" are excluded from score calculations
> 
> ## Screenshots
> The fix changes the appearance from an orange warning badge to a grey neutral badge, removing any negative implication for items that cannot be measured.
> 
> Fixes #618

image

joanagmaia:
Hi @balaraj74 , thank you so much for your interest in contributing, we really appreciate it!

For UI changes, we typically loop in @jonathimer  and @nunoeufrasio to ensure we’re aligned on the intended direction before making updates. The original issue suggests adjusting the color of one of the security tags, but we’d like to take a step back and have a broader discussion to clarify the meaning and improve the overall user experience. Until that conversation takes place, it’s a bit early to move forward with implementation.

With that in mind, I’m going to close this PR for now. Once we kick off the follow-up discussion in the original issue, we’d love to revisit and continue from there.

balaraj74:
> Hi @balaraj74 , thank you so much for your interest in contributing, we really appreciate it!
> 
> For UI changes, we typically loop in @jonathimer  and @nunoeufrasio to ensure we’re aligned on the intended direction before making updates. The original issue suggests adjusting the color of one of the security tags, but we’d like to take a step back and have a broader discussion to clarify the meaning and improve the overall user experience. Until that conversation takes place, it’s a bit early to move forward with implementation.
> 
> With that in mind, I’m going to close this PR for now. Once we kick off the follow-up discussion in the original issue, we’d love to revisit and continue from there.

Thanks for reviewing. And can you please tell how can I continue this

jonathimer:
I do like the idea, but we should make sure we pick a color matching our UI style cc @nunoeufrasio

PR #1453, opened 2025-11-28
feat: use Kustomize for deploying to K8s

borfast:
This introduces [Kustomize](https://kustomize.io/) to deploy Insights and at the same time uses OCI secrets instead of the environment variables from crowd-kube.

Kustomize is kind of like Helm but has a bunch of advantages, including being natively supported in kubectl.

This would allow us to get rid of the various prod/staging folders in crows-kube, and instead have a single set of files for both staging and production - and even local development, if we ever wanted to start using something like minikube or kind for local development.

My intention was to do it first on Insights because it's a simpler project, and then eventually we could also apply it to the crowd.dev repository. Maybe someone wants to pick it up where I left off.

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,0 +1,6 @@
+{
+  "sandbox": {
+    "enabled": true,
+    "autoAllowBashIfSandboxed": true
+  }
+}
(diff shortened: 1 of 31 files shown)


joanagmaia:
Closing this PR until the team finds some time to work on the migration of environment variables to OCI secrets.
Won't delete branch or PR as we can use some of this implementation in the future

PR #1575, opened 2026-01-13
fix runtime error due to missing getHeader

kamilansri:
- `getHeader` is used but NOT imported

`import { getCookie } from 'h3';`

 `getHeader(event, 'user-agent')`
  `getHeader(event, 'referer')`

---

- This will throw `runtime error`

`ReferenceError: getHeader is not defined`


gaspergrom:
Hey, we are not open yet for contributions as explained in readme.
https://github.com/linuxfoundation/insights/blob/main/README.md#contributing

kamilansri:
It would be a great opportunity of contribution , when you are open.
Thanks @gaspergrom ,

PR #1691, opened 2026-02-19
fix: copilot by using bucketId and adjust agent models

joanagmaia:
This pull request introduces significant improvements to the Data Copilot's agent orchestration and data auditing, focusing on model selection, Tinybird integration, and auditor prompt accuracy. The changes enable more precise use of language models for different agents, ensure correct data partitioning with Tinybird's `bucketId`, and enhance the auditor's ability to provide user-relevant summaries by including actual data samples.

**Model selection and agent orchestration:**
- Refactored the `DataCopilot` class to use two separate Bedrock models: Sonnet for routing and pipe agents, and Opus for text-to-SQL and auditor agents, allowing for more optimal model usage per agent type.

**Tinybird integration and data partitioning:**
- Added logic to fetch and cache the Tinybird `bucketId` per project, injecting it into all relevant Tinybird pipe calls and tool executions to ensure correct data partitioning and prevent cross-project data leakage.

**Auditor prompt and data summary improvements:**
- Enhanced the auditor prompt to include the top rows of actual data (not just statistics), requiring the summary to reference these real values and handle unknown or placeholder entries appropriately.
- Adjusted the auditor output schema to allow `feedback_to_router` and `summary` to be nullable, reflecting cases where these may not be present.

**Other improvements:**
- Filtered out Tinybird tools with empty descriptions to prevent Bedrock validation errors.
- Minor bug fix: ensured `previousFeedback` is set to `undefined` if not present, improving retry logic robustness.

These changes collectively improve the reliability, accuracy, and user-friendliness of the Data Copilot's responses and its integration with the Tinybird data backend.


joanagmaia:
@epipav can you check it out and let me know what you think? — Summary of the changes in the PR description.

The copilot was quite broken in production, so I made a set of changes to address the main issues.

The primary problem was that `bucketId` logic was never implemented. As a result, most requests were returning `0` because the system couldn’t properly resolve projects (this is the current production behavior). This PR introduces the missing `bucketId` handling to ensure queries are correctly scoped.

I also evaluated using Opus 4.6 across all agents, but it proved too slow—particularly for the Router and Pipe agents, where it would often stall or take too long to respond. Given the current architecture, it doesn’t seem viable to use it there without more significant changes, so I’m now using one model for some agents, and opus model for others.

Additionally, I improved the data summary returned to the user by including the top 3 rows. This increases token usage slightly, but the improvement in response clarity and UX is noticeable. While this won’t cover every edge case, it works well for common queries like “most active X” and similar straightforward requests.

## Let me know if anything here doesn’t look right—especially since you have more context on some parts of the system.

Some tests:
Screenshot 2026-02-25 at 18 08 04
Screenshot 2026-02-25 at 18 14 30
Screenshot 2026-02-25 at 18 12 00

frontend/lib/chat/data-copilot.ts
```diff
@@ -170,10 +214,20 @@ export class DataCopilot {
     instructions?: string,
   ): Promise<void> {
     const chatRepo = new ChatRepository(insightsDbPool);
+
+    let model: string | undefined = this.BEDROCK_SONNET_MODEL_ID;
+
+    if (agent === 'EXECUTE_INSTRUCTIONS') {
+      model = undefined;
+      // For now, we use the Opus model for text-to-SQL and auditor.
+      // The model is currently too slow for both the pipe and router agents.
```
epipav: readability NIT: This comment block should come before `let model: string | undefined = this.BEDROCK_SONNET_MODEL_ID`

frontend/lib/chat/data-copilot.ts
```diff
@@ -56,14 +57,23 @@ export class DataCopilot {
...
 
+  /** Tinybird bucketId for the current project — fetched once and reused across all pipe calls */
+  private bucketId: number | null = null;
+
   /** Tinybird MCP server URL */
   private tbMcpUrl: string = '';
 
-  /** Amazon Bedrock language model instance */
-  private model: LanguageModelV1;
+  /** Amazon Bedrock language model instance for routing and auditing (Sonnet) */
```
epipav: Auditor uses Opus, not Sonnet

frontend/lib/chat/data-copilot.ts
```diff
@@ -56,14 +57,23 @@ export class DataCopilot {
...
+  private sonnetModel: LanguageModelV1;
+
+  /** Amazon Bedrock language model instance for text-to-SQL, pipe, and chart agents (Opus) */
```
epipav: docs look wrong: pipe agent uses Sonnet, not Opus

frontend/lib/chat/utils/data-summary.ts
```diff
@@ -32,8 +33,8 @@ export interface ColumnStats {
 
 /**
  * Generate statistical summary of dataset
- * Token-efficient: ~400-500 tokens for typical dataset
- * No raw data samples sent to LLM - only statistics
+ * Token-efficient: ~1500-200 tokens for typical dataset
```
epipav: There's a typo here, should be 1500-2000

frontend/lib/chat/instructions.ts
```diff
@@ -44,14 +44,19 @@ async function executeTinybirdPipe(
...
+): Promise<any[]> {
   // Execute the pipes according to the instructions
   const pipeResults: Record<string, any[]> = {};
 
   // Execute each pipe with its inputs using TinyBird API
   for (const pipeInstruction of instructions.pipes) {
     try {
-      const result = await executeTinybirdPipe(pipeInstruction.name, pipeInstruction.inputs);
+      const inputs =
+        bucketId !== null ? { bucketId, ...pipeInstruction.inputs } : pipeInstruction.inputs;
```
epipav: The optional `bucketId` param can also be undefined here when the function call doesn't have `bucketId`. Checking with loose equality (`bucketId != null`) also catches undefined here

frontend/lib/chat/data-copilot.ts
```diff
@@ -88,10 +99,43 @@ export class DataCopilot {
...
+      process.env.NUXT_TINYBIRD_BASE_URL || 'https://api.us-west-2.aws.tinybird.co';
+    const tinybirdToken = process.env.NUXT_INSIGHTS_DATA_COPILOT_TINYBIRD_TOKEN;
+    if (!tinybirdToken) return;
+    try {
+      const response = await ofetch(
+        `${tinybirdBaseUrl}/v0/pipes/project_buckets.json?project=${encodeURIComponent(project)}`,
+        { headers: { Authorization: `Bearer ${tinybirdToken}` }, timeout: 10_000 },
+      );
+      this.bucketId = response.data?.[0]?.bucketId ?? null;
+      console.warn(`🪣 [DataCopilot] bucketId for "${project}": ${this.bucketId}`);
```
epipav: NIT: let's use `console.log` here, since it's a non-warning log

frontend/lib/chat/utils/data-summary.ts
```diff
@@ -127,9 +129,13 @@ export function generateDataSummary<T extends Record<string, unknown>>(data: T[]
     columnStats[col] = stats;
   }
 
+  const rows = data as Record<string, unknown>[];
   return {
     rowCount: data.length,
     columns,
     columnStats,
+    // Ideally the entire response would be available to the auditor. But that would be too costly.
+    // TODO: Explore a better way to have a proper summary of the data that answers the user's question directly.
```
epipav: NIT: This TODO comment doesn't add much here - Data summary already tries to tackle this somewhat, and if it needs better summarization, I think we can create a task for it

frontend/lib/chat/data-copilot.ts
```diff
@@ -88,10 +99,43 @@ export class DataCopilot {
       }),
     });
 
-    this.tbTools = await this.mcpClient.tools({});
+    const allTools = await this.mcpClient.tools({});
+
+    // Filter out tools with empty descriptions — Bedrock rejects them with a validation error
+    this.tbTools = Object.fromEntries(
```
epipav, line 105: Anything important is missing descriptions? Should we report these to Tinybird support?

PR #1709, opened 2026-02-26
feat: collections redesign

emlimlf:
## In This PR

- Redesigned the featured collections in the landing page and removed the carousel
- Added new routes for curated, community and my collections
- Re-designed the collections list page to match new designs
- Re-designed the collections details page to match new designs

Note: All details related to the community and my collections as well as the like and create collections are hidden or not included in this PR

## Ticket

[IN-984](https://linuxfoundation.atlassian.net/browse/IN-984)

[IN-984]: https://linuxfoundation.atlassian.net/browse/IN-984?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

- `main` 
  - \#1709 :point\_left:
    - \#1714

PR #1714, opened 2026-02-27
feat: create collection

emlimlf:
## In this PR

- Added the UI for the create collections. It includes the modal and all the steps
- Added check for Insights team member so that the create collection and the other tabs will only appear for specific users only

Note: This is just a UI and does not save data. This PR: <https://github.com/CrowdDotDev/crowd-kube/pull/86> is also required to be pushed to prod for this to work.

## Ticket

[IN-987](https://linuxfoundation.atlassian.net/browse/IN-987)

[IN-987]: https://linuxfoundation.atlassian.net/browse/IN-987?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

- `main` 
  - \#1709
    - \#1714 :point\_left:


frontend/app/components/modules/collection/components/create-modal/modal-header.vue
```diff
@@ -0,0 +1,82 @@
...
+            name="rectangle-history-circle-plus"
+            :size="24"
+            class="col-1 row-1 text-accent-500 m-auto"
+          />
+        </div>
+        <!-- Title and description -->
+        <div class="flex flex-col gap-1">
+          <h2 class="text-xl font-bold font-secondary leading-7 text-neutral-900">Create collection</h2>
+          <p class="text-xs font-normal leading-4 text-neutral-500">
+            Create you own collection to organize or showcase open-source projects around a specific theme or purpose.
```
gaspergrom: Small typo: "Create you own collection" should be "Create **your** own collection".

frontend/app/components/modules/collection/components/create-modal/create-collection-modal.vue
```diff
@@ -0,0 +1,129 @@
...
+}>();
+
+const isModalOpen = computed({
+  get: () => props.modelValue,
+  set: (value: boolean) => emit('update:modelValue', value),
+});
+
+const step = ref(0);
+const form = ref<CreateCollectionForm>({ ...createCollectionTemplate });
+const stepRef = ref<{ $v?: { $invalid: boolean; $touch: () => void } } | null>(null);
```
gaspergrom, line 74: Heads up — the spread here does a shallow copy, so `form.value.projects` and `createCollectionTemplate.projects` point to the same array reference. Right now it's fine because `step-projects.vue` always reassigns the array (no `.push()` mutations), but it's fragile. If someone later does `form.value.projects.push(...)` it would also mutate the template.

Worth making this defensive:
```ts
const form = ref<CreateCollectionForm>({ ...createCollectionTemplate, projects: [] });
```

Same applies to the reset in the `watch` and `onMounted` below.

frontend/app/components/modules/collection/components/create-modal/create-collection-modal.vue
```diff
@@ -0,0 +1,129 @@
...
+});
+
+onMounted(() => {
+  form.value = { ...createCollectionTemplate };
+});
+</script>
+
+<script lang="ts">
+export default {
+  name: 'LfCreateCollectionModal',
```
gaspergrom, line 123: This `onMounted` is redundant — `form` is already initialized with `{ ...createCollectionTemplate }` on line 74. You can drop this entirely.

frontend/server/utils/jwt.ts
```diff
@@ -65,3 +65,9 @@ export const hasLfxInsightsPermission = (claims?: string[]): boolean => {
 
   return claims.includes(config.lfxAuth0TokenClaimGroupName);
 };
+
+export const isLfInsightsTeamMember = (email: string): boolean => {
+  const config = useRuntimeConfig();
+  const lfxInsightsTeam = config.lfxInsightsTeam || '';
```
gaspergrom: If the env var has any spaces (e.g. `"alice@lf.org, bob@lf.org"`), the `.includes()` check will fail because `.split(',')` gives `["alice@lf.org", " bob@lf.org"]`.

Safe to add a `.map(e => e.trim())`:
```ts
return lfxInsightsTeam.split(',').map((e) => e.trim()).includes(email);
```

frontend/app/components/modules/collection/config/collection-type-config.ts
```diff
@@ -4,41 +4,55 @@
...
-  //   detailsLabel: 'My Collections',
-  //   icon: 'folder-heart',
-  //   route: LfxRoutes.COLLECTIONS_MY_COLLECTIONS,
-  //   activeClass: '!bg-discovery-200',
-  //   iconHighlightClass: '!bg-discovery-500',
-  //   type: 'my-collections',
-  // },
-];
+export const collectionTabs = (user: User | null): CollectionTypesTabs[] => {
+  return !!user?.isLfInsightsTeamMember
```
gaspergrom: The curated tab config is duplicated in both branches of the ternary. Could simplify by always including curated and conditionally appending the rest:
```ts
export const collectionTabs = (user: User | null): CollectionTypesTabs[] => {
  const tabs: CollectionTypesTabs[] = [
    {
      label: 'Curated',
      detailsLabel: 'Curated Collections',
      icon: 'gem',
      route: LfxRoutes.COLLECTIONS_CURATED,
      activeClass: '!bg-neutral-200',
      iconHighlightClass: '!bg-neutral-900',
      type: 'curated',
    },
  ];

  if (user?.isLfInsightsTeamMember) {
    tabs.push(
      { /* Community */ },
      { /* My Collections */ },
    );
  }

  return tabs;
};
```
Easier to maintain and less chance of the two curated configs drifting apart.

PR #1724, opened 2026-03-05
feat: collections discovery UI

emlimlf:
## In this PR

- Added new UI for collections discovery using mock data
- Re-enabled the collections menu on the site header

## Ticket
[IN-988](https://linuxfoundation.atlassian.net/browse/IN-988)

[IN-988]: https://linuxfoundation.atlassian.net/browse/IN-988?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


frontend/app/components/modules/collection/config/collection-type-config.ts
```diff
@@ -59,3 +72,37 @@ export const headerBackground = (type?: CollectionType) => {
...
+    label: 'My Collections',
+    detailsLabel: 'My Collections',
+    description: "Collections you've created or liked.",
+    icon: 'folder-heart',
+    route: LfxRoutes.COLLECTIONS_MY_COLLECTIONS,
+    activeClass: '!bg-discovery-200',
+    iconHighlightClass: '!bg-discovery-500',
+    type: 'my-collections',
+  },
+];
```
gaspergrom: The tab data (labels, icons, routes, descriptions) is now duplicated between `collectionTabs()` and `discoveryCollectionTabs`. If someone updates the description or adds a new tab, they'd have to remember to change both. Could we have `discoveryCollectionTabs` be the single source of truth and just filter it inside `collectionTabs()` based on the user?

frontend/app/components/shared/components/collection-section.vue
```diff
@@ -0,0 +1,89 @@
...
+  },
+);
+
+const title = computed(() => discoveryCollectionTabs.find((tab) => tab.type === props.type)?.detailsLabel || '');
+const subtitle = computed(() => discoveryCollectionTabs.find((tab) => tab.type === props.type)?.description || '');
+const icon = computed(() => discoveryCollectionTabs.find((tab) => tab.type === props.type)?.icon || '');
+const iconBackground = computed(
+  () => discoveryCollectionTabs.find((tab) => tab.type === props.type)?.iconHighlightClass || '',
+);
+const viewAllRoute = computed(() => discoveryCollectionTabs.find((tab) => tab.type === props.type)?.route || '');
```
gaspergrom: We're calling `.find()` on the same array with the same predicate 5 times for each render. Minor thing, but it'd be cleaner to compute the matched tab once and derive the individual properties from it:
```ts
const tab = computed(() => discoveryCollectionTabs.find((t) => t.type === props.type));
const title = computed(() => tab.value?.detailsLabel || '');
// etc.
```

frontend/app/components/modules/collection/components/create-modal/create-button.vue
```diff
@@ -0,0 +1,39 @@
...
+    type="secondary"
+    button-style="pill"
+    @click="isCreateCollectionModalOpen = true"
+  >
+    <lfx-icon name="rectangle-history-circle-plus" />
+    Create collection
+  </lfx-button>
+
+  <lf-create-collection-modal
+    v-if="isCreateCollectionModalOpen"
```
gaspergrom, line 17: The `v-model` already handles two-way binding, so the explicit `@update:model-value` handler here is redundant. You can safely remove the `@update:model-value` line.

frontend/app/components/modules/collection/views/collection-discover.vue
```diff
@@ -0,0 +1,146 @@
...
+const {
+  data: discoveryData,
+  status: discoveryStatus,
+  error: discoveryError,
+  suspense: discoverySuspense,
+} = COLLECTIONS_API_SERVICE.fetchDiscoveryCollections();
+
+const curatedCollections = computed(() => discoveryData.value?.curatedCollections || []);
+const communityCollections = computed(() => discoveryData.value?.communityCollections || []);
+const myCollections = computed(() => discoveryData.value?.myCollections || []);
```
gaspergrom, line 125: These `as unknown as Record<string, unknown>[]` casts are a code smell — `isEmptyData` should probably accept `Collection[]` directly (or at minimum an `unknown[]`). Worth a quick fix to the helper signature rather than triple-casting everywhere.

frontend/app/components/modules/collection/views/collection-discover.vue
```diff
@@ -0,0 +1,146 @@
...
+            />
+          </div>
+        </section>
+      </template>
+    </div>
+  </div>
+</template>
+
+<script setup lang="ts">
+import { computed, onServerPrefetch } from 'vue';
```
gaspergrom, line 100: Heads up — the `icon-style="solid"` prop here is different from the `type="solid"` pattern used elsewhere (e.g. in `collection-card.vue` line 157). Not sure if both work, but worth checking for consistency.

frontend/docs/images/repository-groups.png
gaspergrom, line 1: what is this image here

emlimlf: I don't know, I've never touched the docs section

frontend/app/components/modules/collection/config/collection-type-config.ts
emlimlf: done

frontend/app/components/modules/collection/components/create-modal/create-button.vue
emlimlf: done

frontend/app/components/modules/collection/views/collection-discover.vue
emlimlf: done

PR #1833, opened 2026-04-10
feat: sign in wall for collections

gaspergrom:


frontend/app/pages/collection/my-collections/index.vue
```diff
@@ -6,12 +6,18 @@ SPDX-License-Identifier: MIT
   <div class="bg-white lg:!pb-30 pb-20 lg:!-mb-30 -mb-20 flex-grow">
     <lfx-collection-list-view type="my-collections" />
   </div>
+
+  <lfx-collection-auth-wall
+    v-if="showAuthWall"
+    v-model="showAuthWall"
+  />
```
joanagmaia: The background page of the sign in auth wall should not be in a loading state. We are making unnecessary requests which will simply return 401.

Expected behaviour (confirm with Nuno): User is never able to access the “My Collections“ page. As soon as they click on the “My collections” tab we show the modal in the page we were before. If the user tries to access the page from the URL directly, we send them straight to authentication without ever showing the tab.

joanagmaia (changes requested):

joanagmaia (approved):
LGTM

PR #1842, opened 2026-04-15
feat: add unavaliable health score state

gaspergrom:


joanagmaia (approved):
Looks good! I would still apply the suggestions from copilot on the `!score` validation which could fail if score was actually 0. We should probably check against null value or whatever we return when score is unavailable?

PR #1926, opened 2026-05-19
docs: add instructions to update submodules

gbrsni:
The existing instructions put the user in a state in which `pnpm install` would fail due to the missing submodules. I added instructions to clone the repo with submodules and to update submodules in case the user followed the original instructions.


joanagmaia (approved):
LGTM, thanks for the contribution!

PR #2089, opened 2026-08-14
fix: stabilize OG image generation under bot crawler load

gaspergrom:
## Summary

Implements multi-layered approach to fix OG image 500 errors under concurrent bot crawler load (IN-1200):

- **Timeout handling** (8s): Prevents slow Satori/Resvg rendering from blocking
- **Bot rate limiting** (5 req/s per bot): Targets bingbot, PetalBot, Amzn-SearchBot
- **Response caching** (24h): Reduces rendering load from repeated requests
- **Collection validation**: Mirrors project middleware, prevents invalid data to renderer
- **Enhanced error handling**: Better logging and timeout detection

## Files Changed

- `frontend/server/middleware/og-image-timeout.ts` (new)
- `frontend/server/middleware/og-image-bot-limiter.ts` (new)
- `frontend/server/middleware/og-image-cache-headers.ts` (new)
- `frontend/server/middleware/og-image-collection.ts` (new)
- `frontend/server/plugins/og-image-fallback.ts` (enhanced)
- `frontend/setup/og-image.ts` (cache config added)

## How to Test

1. Verify timeout kicks in for slow renders
2. Test rate limiting with bot user agents
3. Confirm cache headers are set on responses
4. Validate collection endpoint redirects deleted collections to fallback
5. Load test with concurrent bot-like requests

## Out of Scope

- robots.txt modifications
- CDN-level edge caching
- Kubernetes/ingress bot throttling
- Rendering service resource tuning

These infrastructure changes are being tracked separately.

PR #2158, opened 2026-09-04
fix: pin h3 dependency to resolve OG image 500 errors IN-1275

gaspergrom:
## Summary

- Fixes production 500s on `/__og-image__/image/project/:slug/...` for real project slugs carrying certain query params (confirmed repro: HubSpot marketing tracking params like `__hstc`/`__hssc`/`__hsfp`).
- Root cause: `nuxt-og-image@5.1.13`'s compiled code imports `getQuery`/`createError` from `h3` without declaring `h3` as its own dependency. pnpm had no fixed slot for that import and resolved it to an incompatible `h3@2.0.1-rc.16` (pulled in via a stale duplicate `nuxt-site-config@3.2.11`) instead of the app's `h3@1.15.5`. h3 v2's `getQuery` does `new URL(event.req.url)`, which throws `TypeError: Invalid URL` for a relative `req.url` — i.e. every real Nitro request.
- An existing Nitro error-hook fallback (`server/plugins/og-image-fallback.ts`, from IN-1200/PR #2089) is meant to catch exactly this and 302-redirect instead, but it could lose a race against nitropack's default error handler under production timing, sometimes surfacing the raw 500 instead.
- Fix is dependency-resolution only, in `pnpm-workspace.yaml`: a `packageExtensions` entry gives `nuxt-og-image` its own explicit `h3: 1.15.5` slot, plus a scoped `'nuxt-og-image>nuxt-site-config': ^3.2.18'` override to collapse the stale duplicate. No application code changed.
- Side effect: collapsing that duplicate shifted vitest's peer-resolved `vite` down to a version vitest 4 rejects. Added a `vite: 7.3.1` override to match — this consolidates `main`'s 3 duplicate vite installs into 1, not a downgrade.
- **Blast-radius scan (post-fix):** the same phantom-import pattern — importing from `h3` without declaring it as a dependency — turned out not to be unique to `nuxt-og-image`. Checked every other Nuxt-SEO-ecosystem package in the tree:
  - `@nuxtjs/sitemap@7.5.2` has the identical gap and was resolving `h3` the same unsafe way, just not yet observed crashing in production. Fixed with the same `packageExtensions` pattern.
  - `nuxt-schema-org@5.0.10`, `nuxt-link-checker@4.3.9`, `nuxt-seo-utils@7.0.19` have the same undeclared-dependency gap (different h3 functions than the crashing ones, so not confirmed exploitable) — pinned preemptively rather than leaving the same latent bug class for a future ticket.
- Does not change OG-image rendering logic, the existing fallback plugin, or any other route/feature.

## Changes

| File | What changed |
|------|-------------|
| `pnpm-workspace.yaml` | `packageExtensions` entries pinning the phantom `h3` import to `1.15.5` for `nuxt-og-image`, `@nuxtjs/sitemap`, `nuxt-schema-org`, `nuxt-link-checker`, and `nuxt-seo-utils`; `nuxt-og-image>nuxt-site-config` override to collapse the stale duplicate; `vite: 7.3.1` override to fix the resulting vitest peer-resolution break |
| `frontend/server/plugins/og-image-h3-resolution.test.ts` | New regression test — resolves `h3` the same way `nuxt-og-image`'s compiled code does and asserts `getQuery` doesn't throw on the ticket's repro request shape (relative `req.url` with HubSpot tracking params) |
| `pnpm-lock.yaml` | Regenerated lockfile reflecting the above resolution changes |

## JIRA

IN-1275 — Investigate 500 Internal Server Error on insights.linuxfoundation.org OG image generation

## Deploy order

_No deploy ordering constraints — self-contained to this repo. Dependency-resolution fix only; no upstream/downstream contract changes._

## DB migrations

_No DB migrations._

## Test plan

- [x] `pnpm test` — 251/251 pass, including the new regression test
- [x] Manual curl repro under `pnpm dev`: plain project OG-image → 200
- [x] Manual curl repro under `pnpm dev`: project + HubSpot tracking params (`__hstc`/`__hssc`/`__hsfp`, the exact IN-1275 repro shape) → 200
- [x] Manual curl repro under `pnpm dev`: non-existent slug → 302 fallback (matches pre-existing IN-1200 behavior, no regression)
- [x] Verified against a fresh install (`node_modules` wiped, reinstalled from scratch) — same clean resolution, same 200s
- [x] 11-case edge-case matrix run against the route (malformed query chars, broken JSON, duplicate keys, oversized query string, empty query, multiple crawler user agents, collection-scoped og-image route) — zero 500s
- [ ] `pnpm build` against the real production Nitro bundle — not verified locally; blocked by an unrelated pre-existing `@resvg/resvg-js-win32-x64-msvc` ENOENT (a Windows-only native binary lookup) on this macOS dev machine. Verify in CI/staging before merge.

Note: `pnpm lint`/`lint:fix` fail identically on this branch and on unmodified `main`, due to an unrelated pre-existing `eslint-plugin-vuejs-accessibility`/`globals` phantom-dependency issue — not introduced or affected by this PR.

## Checklist

- [x] `git commit --signoff -S` on every commit
- [x] MIT license header on every new source file
- [x] PR diff < 1000 lines (mostly lockfile)
- [x] No unrelated changes bundled — the 3-package preemptive fix is the same root-cause bug class as the ticket, not scope creep
- [x] Cross-repo impact documented (none — dependency-resolution fix contained to this repo)

### DIFF
--- a/frontend/server/plugins/og-image-h3-resolution.test.ts
+++ b/frontend/server/plugins/og-image-h3-resolution.test.ts
@@ -1,0 +1,41 @@
+// Copyright (c) 2025 The Linux Foundation and each contributor.
+// SPDX-License-Identifier: MIT
+
+import { createRequire } from 'node:module';
+import { describe, it, expect } from 'vitest';
+
+// IN-1275: nuxt-og-image's compiled code imports `getQuery`/`createError` from `h3`
+// without declaring `h3` as its own package.json dependency. pnpm therefore has no
+// fixed slot to resolve that import into, and (absent the packageExtensions/overrides
+// fix in pnpm-workspace.yaml) can resolve it to h3 v2 instead of the app's h3 v1.
+// h3 v2's getQuery does `new URL(event.req.url)`, which throws `TypeError: Invalid URL`
+// for a real Node request whose `req.url` is a relative path (e.g. an OG-image request
+// carrying HubSpot tracking params like __hstc/__hssc/__hsfp) — this was the exact
+// cause of the raw 500s reported in IN-1275. h3 v1's getQuery has no such requirement.
+//
+// This test resolves `h3` the same way nuxt-og-image's own code resolves it (Node
+// module resolution from nuxt-og-image's package), so it reflects whatever pnpm
+// actually links today rather than asserting a version number.
+describe('nuxt-og-image h3 resolution (IN-1275)', () => {
+  it('does not throw when reading query params from a real OG-image request', async () => {
+    const nuxtOgImageEntry = import.meta.resolve('nuxt-og-image');
+    const requireFromNuxtOgImage = createRequire(nuxtOgImageEntry);
+    const h3 = requireFromNuxtOgImage(requireFromNuxtOgImage.resolve('h3'));
+
+    // Shape of a real Node request as seen by Nitro: `url` is a relative path,
+    // not an absolute URL.
+    const nodeReq = {
+      url: '/__og-image__/image/project/test-project/og.png?__hstc=1&__hssc=2&__hsfp=3',
+      method: 'GET',
+      headers: {},
+    };
+    const nodeRes = { getHeader() {}, setHeader() {}, end() {}, writeHead() {}, once() {} };
+
+    // fails before fix: with h3 v2 resolved, constructing the event (or calling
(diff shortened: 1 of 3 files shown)


gaspergrom:
Closing because this is fixed by the upgrade

PR #2174, opened 2026-09-09
docs: relocate public API architecture docs under api/docs/arch/

epipav:
## Summary

- Move the public API architecture plan, ADRs, and architecture review from `docs/` to `api/docs/arch/`, colocating them with the `api/` package they describe.
- Fix the two stale path references inside `architecture-review/01-overview.md` (ADR link, `libs/types` name) to match the new location and actual lib name.

## Test plan

- [x] Diff is rename-only aside from the two path-reference line fixes noted above

* `main` 
  - **docs: relocate public API architecture docs under api/docs/arch/** :point\_left:
    - \#2175
      - \#2176
        - \#2177


gaspergrom (approved):
Looks good to me, LGTM.

PR #2175, opened 2026-09-09
docs: align ADR rule and skill with api/docs/arch/adr path

epipav:
## Summary

- Update `.claude/rules/adr-format.md` and `.claude/skills/adr/SKILL.md` to reference `api/docs/arch/adr/` instead of the old `docs/adr/`.
- Also folds in the ADR template guidance alignment (free-form prose vs. full template, exemption for README/template.md) so the skill matches the rule exactly.

## Test plan

- [x] Rule/skill content reviewed for consistency with the relocated path

* `main` 
  - \#2174
    - **docs: align ADR rule and skill with api/docs/arch/adr path** :point\_left:
      - \#2176
        - \#2177

### DIFF
--- a/.claude/rules/adr-format.md
+++ b/.claude/rules/adr-format.md
@@ -1,12 +1,12 @@
 ---
-description: Enforce ADR structure and conventions on files written under docs/adr/
+description: Enforce ADR structure and conventions on files written under api/docs/arch/adr/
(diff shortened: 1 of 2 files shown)


gaspergrom (approved):
LGTM, matches the new path everywhere.

PR #2201, opened 2026-09-14
feat: wire openapi export script

epipav:
## Summary

Adds a release-time OpenAPI export to the API package (IN-1133, part of the IN-1130 sequence: IN-1131 → IN-1133 → IN-1132 → IN-1134/IN-1135).

The Fastify app already registers `@fastify/swagger` and serves the generated document live at `/v1/openapi.json`. This adds `pnpm openapi:export`, a script that builds the same app, waits for it to be ready, and writes the spec to a file without ever binding a port — so the docs build (IN-1132) can ingest an exact file artifact on every release.

## What changed

- `api/scripts/export-openapi.ts` — new export script. Defaults to `api/openapi/v1.json`, accepts an optional CLI arg to override the path, creates missing directories, exits non-zero on failure.
- `api/package.json` — new `openapi:export` script; widened `lint`/`format` globs to also cover `scripts/`.
- `api/.gitignore` — `openapi/` (release artifact, regenerated every release).
- `api/tsconfig.json` — `scripts/**/*.ts` added to `include` so the script typechecks.
- `api/tests/openapi-export.test.ts` — integration-tier tests covering default path, CLI override, structural OpenAPI 3.x validity, parity with the live `/v1/openapi.json` endpoint, and non-zero exit on write failure.

## How to test

```
cd api
pnpm openapi:export            # writes api/openapi/v1.json
cat api/openapi/v1.json        # inspect the generated spec
pnpm openapi:export /tmp/spec.json   # custom path
pnpm test                      # full suite green
```

## Notes for reviewers

Non-blocking follow-ups noted during review, left for later passes:
- The exported document bakes in whatever `API_PUBLIC_URL` (or its `localhost:4000` default) was set when the export ran, since it reuses the runtime `servers` config. Fine for local/dev use; CI/release wiring for this should set the real public URL when IN-1132 lands.
- A couple of test-suite ergonomics items (the default-path test clears `api/openapi/` in the working tree rather than a temp dir; two tests assert literal `package.json`/`.gitignore` strings rather than behavior) are worth cleaning up in a follow-up, not blocking this change.


joanagmaia (approved):
LGTM 🚀

PR #2208, opened 2026-09-15
fix: coerce githubAppId to a string before signing the jwt

epipav:
## Summary
- Fixes a production 500 on `POST /api/report`: `TypeError: "iss" claim must be a string`.
- `getInstallationToken()` passed `githubAppId` straight into jose's `SignJWT.setIssuer()`. Nuxt's runtime config coerces numeric-looking env vars (e.g. `NUXT_GITHUB_APP_ID`) into JS numbers, so `githubAppId` arrives as a `number`, not a `string`.
- Fix: coerce with `String(githubAppId)` before signing. Extracted the signing logic into a standalone `signAppJwt()` so this exact seam has a direct regression test (Nuxt's `useRuntimeConfig`/`#imports` auto-import machinery isn't mockable in this repo's Nuxt/Vitest setup, so testing through the original caller wasn't practical).
- Also fixes `pnpm-workspace.yaml`'s `supportedArchitectures` (adds `musl` alongside `glibc`) — needed to resolve native bindings for `oxc-parser`/`rolldown` in this environment without breaking the musl/Alpine-based worker Docker builds.

## Why this happened
The bug was latent since #2046 ("feat: mirror report issues to jira", merged 2026-08-10), which introduced the untyped `.setIssuer(githubAppId)` call. It only became a live 500 after #2159 ("chore: upgrade dependencies", merged 2026-09-07, jose 6.1.3 → 6.2.12), which added strict runtime validation requiring the JWT `iss` claim to be a string.

## Test plan
- [x] `signAppJwt` regression tests cover `githubAppId` as both `number` and `string`, asserting the decoded `iss` claim is always a string.
- [x] `pnpm tsc-check` and `pnpm lint` pass.
- [x] Confirmed RED-for-the-right-reason against the original `.setIssuer(githubAppId as string)` (reproduces the exact production error), then GREEN after the fix.


joanagmaia (approved):
LGTM 🚀 Thanks for checking this one out 🙏

PR #2213, opened 2026-09-16
fix: render signal-scores heatmap as a true percentile-band distribution IN-1304

gaspergrom:
## Summary

The "Where the strongest repositories pull ahead" heatmap (added in #2204) only plotted two data points per signal — top-20% and median — recolored as heatmap cells instead of bars. That doesn't deliver an actual percentile distribution, just the same two-point comparison in a different chart type.

This switches the widget to a genuine percentile-band heatmap: 11 signals × 10 columns (10th through 100th percentile, in steps of 10), so each row shows the real spread of repository scores across the whole range.

- `types/report/health-score-coverage-signal-scores.types.ts` — row/data shapes now carry all 10 percentile columns (`p10_pct`..`p100_pct`) instead of `p80_pct`/`median_pct`
- `server/data/tinybird/report/health-score-coverage-signal-scores.ts` — mapper updated to the new shape
- `signal-scores.vue` — heatmap now renders 10 columns with "10th"..."Max" labels; tooltip shows the exact percentile and percentage on hover
- Tests updated to the new row/data shape

## Deploy ordering — read before merging

This depends on a companion pipe change: [linuxfoundation/crowd.dev#4637](https://github.com/linuxfoundation/crowd.dev/pull/4637), which extends `health_score_report_signal_scores` to return the 10 percentile columns instead of `p80_pct`/`median_pct`.

**That pipe has NOT been deployed to production yet** — it's a breaking schema change and deploying it before this frontend PR is live would break the widget silently (rendering `undefined` values with no error). Deploy order must be:

1. Merge and deploy this frontend PR (or have it ready to deploy immediately)
2. Push `crowd.dev#4637`'s pipe to production
3. Confirm the live widget renders correctly against the new pipe output

Do not push the pipe to production ahead of this PR going live, and don't merge this PR expecting it to auto-deploy days before the pipe does — coordinate the two.

## Test plan

- [x] `pnpm lint:fix` — 0 errors (pre-existing unrelated warnings only)
- [x] Unit tests for the mapper updated and passing
- [x] Visually verified the heatmap renders correctly (10 columns, correct labels, working hover tooltips) via a temporary local mock standing in for the not-yet-deployed pipe response — reverted before commit, not part of this diff
- [ ] Final live verification against the real production pipe, once `crowd.dev#4637` deploys


frontend/server/data/tinybird/report/health-score-coverage-signal-scores.ts
```diff
@@ -25,8 +26,12 @@ export function mapHealthScoreCoverageSignalScoreRows(
   const signals: HealthScoreCoverageSignalScore[] = rows.map((row) => ({
     signalKey: row.signal_key,
     categoryKey: row.category_key,
-    p80Pct: row.p80_pct,
-    medianPct: row.median_pct,
+    percentiles: Object.fromEntries(
+      SIGNAL_SCORE_PERCENTILES.map((p) => [
+        p,
+        row[`p${p}_pct` as keyof HealthScoreCoverageSignalScoreRow] as number,
```
themarolt, line 32: per the PR's own deploy order (frontend first, pipe second), this reads the old-shape row until #4637 ships - `row['p10_pct']` etc. don't exist on that row, so 8 of 10 percentiles come back `undefined` and the chart renders `NaN` cells in prod. `/api/report/health-score-coverage/**` is cached in redis for a day, so that stretches past the pipe deploy too unless the cache gets purged. worth making this additive (fall back to the old columns when the new ones are missing) or purging the cache as part of the deploy step - or is the plan to accept a short NaN window and purge manually?

themarolt (approved):

epipav (approved):

PR #2214, opened 2026-09-16
feat: add api version routing structure

epipav:
## Summary

Introduces the URL-prefix API versioning skeleton for the Insights API (parent task: URL-prefix versioning across `/v1-alpha`, `/v1`, `/v2`).

- Each API version is its own Fastify plugin (router tree) under `api/src/versions/<version>/`
- A central registry (`api/src/versions/registry.ts`) enumerates versions with their URL prefixes; `app.ts` registers each plugin under its prefix from that registry
- The existing `GET /v1/openapi.json` handler moved into the v1 tree as a relative route; its public URL is unchanged
- Version plugins are typed `FastifyPluginAsyncTypebox` so TypeBox schemas keep producing typed request/reply inside version trees
- Adding a future version = new plugin directory + one registry entry; version differences are expressed by which routes exist in which tree, never by flags inside handlers

Out of scope (follow-up subtasks): per-version OpenAPI artifacts (IN-1138), Deprecation/Sunset headers (IN-1139), version-bumping playbook (IN-1140).

## Jira

[IN-1137](https://linuxfoundation.atlassian.net/browse/IN-1137) (parent [IN-1136](https://linuxfoundation.atlassian.net/browse/IN-1136))

## Testing

- `api/tests/version-routing.test.ts`: registry shape, registry-driven wiring, `/v1/openapi.json` 200 + OpenAPI 3, scope isolation (root and `/v2` 404)
- `pnpm test` (39 passed), `pnpm tsc-check`, `pnpm lint` all green in `api/`

[IN-1137]: https://linuxfoundation.atlassian.net/browse/IN-1137?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/tests/version-routing.test.ts
```diff
@@ -0,0 +1,76 @@
...
+describe('app wiring (AC1, AC4)', () => {
+  const appSource = readFileSync(`${apiRoot}src/app.ts`, 'utf-8');
+
+  it('app.ts registers versions from the registry, not inline', () => {
+    expect(appSource).toContain('versionRegistry');
+  });
+
+  it('app.ts defines no per-version routes itself', () => {
+    // No hardcoded version-prefixed paths like '/v1/...' outside the registry.
+    expect(appSource).not.toMatch(/['"`]\/v\d/);
```
themarolt: these two assert on the text of `app.ts` rather than on behaviour, and the second one fails on anything that looks like a version literal in that file - a redirect, a `/v1-alpha` comment, a log line. the property you want seems to be "no version-prefixed route exists that the registry didn't put there", which you could get from the routes at runtime - something like comparing `app.printRoutes()` / the `onRoute` hook paths against `versionRegistry.map(e => e.prefix)` after `ready()`. worth it, or do you want the source check as a cheap guardrail against someone adding a route in app.ts?

gaspergrom (approved):
Prefix composition works: /v1/openapi.json still resolves. The isolation tests hit real routes for root and /v2 instead of assuming the shape. Clean split between the registry and the v1 plugin.

themarolt (approved):

api/tests/version-routing.test.ts
epipav: Fixed in `79a82883`: dropped both source checks and replaced them with a runtime test that parses `printRoutes({ commonPrefix: false })` and asserts every route sits under a registry prefix. Also catches version routes added in any other file now :+1:

PR #2215, opened 2026-09-16
fix: address 5xx errors from datadog alerts

gaspergrom:
## Summary

Three unrelated root causes behind the recent Datadog "[Insights] 5xx request response" alert, each fixed independently:

- **`organization-page/*` 500s** — six handlers under `server/api/organization-page/` silently swallowed any non-Tinybird-classified error and rethrew a generic, unlogged 500. Added `logError` before the fallback in all six, matching the existing pattern in `health-score-coverage/bands.get.ts`. No data-side issue was found; the `dash0` contributors data checked out fine independently, so this alert instance was likely transient — the fix makes any recurrence diagnosable.
- **`/api/report` POST 500s** — `createGitHubIssue` had no error handling at all, so any GitHub API failure threw unhandled and unlogged. Wrapped it in try/catch with `logError` and a stable `502` response.
- **`/_ipx/**` 500s for deprecated criticality-badge SVGs** — root cause was not `@nuxt/image` itself (it correctly 404s on a missing source file). The production-only `'**'` catch-all cache route rule in `setup/caching.ts` wraps every unlisted route, including `/_ipx/**`, in Nitro's `cachedEventHandler`, which turns the legitimate 404 into a cached 500 for a full day. Excluded `/_ipx/**` the same way `/_og/**` already is. Verified with a production build before/after.

## Test plan

- [x] `pnpm lint` — 0 errors (only pre-existing `no-explicit-any` warnings, consistent with the established `catch (error: any)` convention used elsewhere)
- [x] `pnpm test` — 255/255 tests pass; 7 pre-existing test-file failures unrelated to this change (confirmed via stash-and-compare — they fail identically without this diff, due to a Tinybird client env-var requirement in the test runner)
- [x] `_ipx` fix verified empirically against a full production build (`NUXT_APP_ENV=production`): missing-file request returns 500 + `cache-control: max-age=86400` before the fix, and a clean 404 with no cache header after; existing images through `_ipx` unaffected


frontend/setup/caching.ts
```diff
@@ -61,6 +61,10 @@ export default {
     '/callback': { redirect: '/api/auth/callback' },
     '/api/auth/**': { prerender: false, index: false, cache: false },
     '/_og/**': { cache: false },
+    // The `/**` catch-all cache rule below wraps every route in Nitro's cachedEventHandler,
```
themarolt, line 64: this new comment says "the `/**` catch-all cache rule below" but the catch-all key is `**`, not `/**` - same slip Copilot flagged on the `/api/security/**` note further down. could we fix the pattern here too since it's new text?

themarolt (approved):

PR #2222, opened 2026-09-17
fix: signal scores heatmap breaks layout on mobile

gaspergrom:
## Summary
- The signal scores heatmap on the health score coverage report ("Where the strongest repositories pull ahead") used a hardcoded 320px label gutter, which left almost no room for the 10 percentile columns on a phone-width card
- Below the `md` breakpoint (768px), the label column shrinks, row labels drop the category prefix (signal name only), and the legend text shortens to fit

## Test plan
- [x] Verified in a 375px-wide viewport: heatmap renders fully, all 10 columns visible, row labels legible with truncation
- [x] `pnpm tsc-check` passes
- [x] `pnpm lint` passes on the changed file
- [x] Desktop layout (≥768px) unchanged — verified against production

IN-1304


themarolt (approved):
lgtm - clean, isolated fix, matches the described mobile viewport bug

PR #2226, opened 2026-09-18
fix: log unclassified errors in project health-score endpoint

gaspergrom:
## Summary

Datadog alert: **[Insights] 5xx request response** (monitor 268253120), two triggering requests:

- `GET /api/project/ClusterDuck/overview/health-score?startDate=2025-09-16&endDate=2026-09-16` → 500
- `GET /api/project/ipaddrjs/overview/health-score?startDate=2025-09-16&endDate=2026-09-16` → 500

Both from a Googlebot user agent. This is a separate endpoint from the one fixed in #2215 (that PR covered `organization-page/*`, `/api/report`, `/_ipx/**`; this alert is on `project/[slug]/overview/health-score`), so it isn't a recurrence of an already-shipped fix.

**Datadog MCP enrichment wasn't available in this session** (not connected), so this was triaged from the pasted alert text alone rather than pulling logs/traces/monitor definition directly from Datadog.

## Investigation

- Confirmed the request's own project-buckets bucket lookup doesn't validate project existence — it deterministically assigns a bucket for any input string, real or not (so a bogus slug doesn't 404 there).
- Queried the underlying Tinybird `health_score_*` pipes directly for both `ClusterDuck` and `ipaddrjs` — all returned clean 200s (empty or populated), no errors.
- Built the app in production mode and curled the **exact** alerted URLs (same slugs, same date params) against real production data. Both returned clean `200`s with fully-populated health-score payloads, including on a repeat request and with date params omitted.
- Tried a genuinely nonexistent slug — correctly returns `404` (not 500), confirming the existing 404-passthrough branch works as intended.
- Could not reproduce the exact alerted failure. This is consistent with a transient upstream/Tinybird blip (matching the pattern already called out in #2215 for a different endpoint), possibly under crawler-driven concurrent load that isn't reproducible from a single request.

Side note (not fixed here, out of scope for this alert): passing an empty project segment (`/api/project//overview/health-score`) triggers a slow Tinybird query that eventually fails — worth a follow-up if it recurs, but unrelated to this alert's real, non-empty slugs.

## Fix

Since the failure didn't reproduce, and this endpoint's catch-all error path was still using an unstructured, mislabeled `console.error('Error fetching active contributors:', error)` (leftover copy-paste from a different handler) instead of the repo's structured `logError` convention already used by the sibling `badge/health-score.ts` handler and the endpoints fixed in #2215 — added `logError` here so any recurrence is diagnosable in Datadog/log aggregation instead of another blind investigation.

## Test plan

- [x] `pnpm exec eslint` on the changed file — no errors
- [x] Production build + local server, curled both exact alerted URLs (200, correct data)
- [x] Repeat request, nonexistent-slug (404), missing-date-params edge cases all verified


themarolt (approved):
lgtm - matches how the other handlers call `logError`, 404 passthrough stays intact

PR #2235, opened 2026-09-21
fix: return 400 instead of 500 for inverted date ranges on widget endpoints

gaspergrom:
## Alert

Datadog monitor **[Insights] 5xx request response** (id `268253120`) fired for:

```
full uri: insights.linuxfoundation.org/api/widget/popularity/package-downloads?project=hermes-agent&granularity=monthly&startDate=2026-09-21&endDate=2025-09-21
method: GET
status: 500
```

Confirmed via the Datadog logs/traces MCP (not just the pasted alert text): the nginx-ingress log for this exact request shows `upstream_status: 500` with a 7-13ms `upstream_response_time` and a `referer` of `.../project/hermes-agent/security?timeRange=custom&start=2026-09-21&end=2025-09-21&generateYaml=true` — a real request with an inverted custom date range (start date a year after end date), not a malformed or bot request.

## Root cause

`getPreviousDates()` in `frontend/server/data/util.ts` throws a plain `Error` when `startDate > endDate`. Nitro has no handler for a plain thrown `Error`, so it surfaces as an uncaught 500 instead of a client error. This function is shared by 17 widget data sources (`popularity`, `development`, `contributors`), so any widget endpoint hit with an inverted date range takes the same path.

## Data validated (Step 3)

`project=hermes-agent` is a real, valid project — the 500 was not caused by bad/missing data. It's the endpoint mishandling a normal (if unusual) input shape: a reversed date range.

## Fix

Changed the plain `throw new Error(...)` to `throw createError({ statusCode: 400, statusMessage: '...' })`, matching the existing pattern already used for this class of case in `frontend/server/data/tinybird/tinybird.ts` (e.g. its 404 "Project not found" throw). This is a one-function fix that applies uniformly to all 17 callers.

## Test plan

- [x] `pnpm exec eslint` on both changed files — clean
- [x] `pnpm tsc-check` — clean (full workspace)
- [x] `pnpm exec vitest run server/data/util.test.ts` — 10/10 passing, including a new regression test asserting `getPreviousDates` throws a `statusCode: 400` error when `startDate > endDate`
- [x] Verified via Datadog that `createError({ statusCode })` from this data layer already correctly surfaces as the given status code end-to-end (the existing 404 case for missing projects does not trigger this monitor's `5xx` query, confirming the pattern works)

## Caveat

This fixes the server returning a 500 for the inverted range. It does not investigate how the `/project/hermes-agent/security` page ended up requesting `start=2026-09-21&end=2025-09-21` (start after end) in the first place — that's a separate, client-side question about the date-range-picker/report-generation flow on that page, out of scope for this alert-driven fix.


themarolt (approved):
lgtm - copilot's point on the two contributors handlers is worth a look before merge

PR #2241, opened 2026-09-21
refactor: register v1-alpha development routes by directory

epipav:
## Why

Every Development endpoint PR (IN-1331 to IN-1335) had to add an import and a `scope.register` line to `api/src/versions/v1-alpha/index.ts`. With five PRs open at once, each merge put the other four into conflict and forced a rebase round. Nine more Development endpoints are still to come, plus the later groups.

## What

- `api/src/versions/v1-alpha/index.ts` loads every module under `development/` with `@fastify/autoload` (`dirNameRoutePrefix: false` because route paths are absolute, `forceESM: true`). An endpoint PR now adds one route file and one test file and touches nothing shared, so merge order stops mattering.
- `api/vitest.config.ts` inlines `@fastify/autoload` (`server.deps.inline`). Vitest transforms `.ts` imports only in project code; autoload's dynamic import runs inside node\_modules and would otherwise hit Node's "Unknown file extension .ts".
- `api/tests/v1-alpha-autoload.test.ts` pins the convention: the modules in `development/` and the `/v1-alpha/projects/{slug}/development/*` paths in the alpha spec match one to one, filename = path leaf. A module that fails to load or registers outside the group now fails a test instead of silently dropping out of the spec.
- New runtime dependency: `@fastify/autoload` 6.5.0 (official Fastify plugin, alongside `@fastify/static` and `@fastify/swagger`).

## Verification

Route tree identical before and after under all three runtimes:

- vitest: 14 files / 196 tests green, plus `tsc-check`, `lint`, `format:check`
- tsx dev server: `/v1-alpha/projects/:slug/development/issues-resolution` registered
- compiled `dist/` under plain Node: same

## Adding an endpoint from now on

Create `api/src/versions/v1-alpha/development/<path-leaf>.ts` with a default-exported Fastify plugin that registers `/projects/:slug/development/<path-leaf>`, and its test under `api/tests/`. Nothing else.

## Follow-up

The four open endpoint PRs (#2238, #2240, #2239, #2237) are rebased onto this branch so their `index.ts` edits disappear; they retarget to `main` when this merges.

Jira: IN-1348

- `main` 
  - **refactor: register v1-alpha development routes by directory** :point\_left:
    - \#2237
    - \#2238
    - \#2239
    - \#2240


gaspergrom (approved):
I checked the autoload config against the manual registration it replaces, and traced through the new test to confirm it asserts a real invariant (filename to route-path-leaf), not something trivially true. Old import and register call are gone. None of the dependent PR's route files show up in this diff. LGTM.

api/src/versions/v1-alpha/index.ts
```diff
@@ -12,7 +13,13 @@ const v1AlphaRoutes: FastifyPluginAsyncTypebox = async (scope) => {
   });
 
   await scope.register(projectRoutes);
-  await scope.register(issuesResolutionRoutes);
+  // Each module under development/ registers its own route, so an endpoint PR adds one file and
+  // never edits this list. Route paths are absolute, hence no directory prefix.
+  await scope.register(autoload, {
```
themarolt, line 16: I think autoload tries to register every script file in the dir, so the first shared helper someone drops next to the routes (`types.ts`, a query builder) fails `app.ready()` instead of being ignored. with 13 more endpoints coming, worth an `ignorePattern` (or a line in docs/arch about where shared code goes), or is the boot failure the point?

themarolt (approved):
lgtm - one question on what happens when a non-route file lands in development/

api/src/versions/v1-alpha/index.ts
epipav: boot failure is the point imo, a stray helper should fail loud instead of an `ignorePattern` quietly skipping a misnamed route module. Documented where shared code goes in `api/README.md` Layout (e9e6b880): `lib/`, `schemas/` or `clients/`.

PR #2246, opened 2026-09-21
feat: add code reviews endpoint under v1-alpha

epipav:
## What

Adds `GET /v1-alpha/projects/{slug}/development/code-reviews`, the `metric=code-reviews` branch of the Nuxt
`code-review-engagement` widget handler, as its own endpoint in the public API.

It returns:

- `summary`: code reviews in the current period against the comparison period before it (`current`, `previous`,
  signed `percentageChange`, `changeValue`, `periodFrom`, `periodTo`), counts as integers.
- `data`: one `{ startDate, endDate, reviews }` per granularity bucket, ISO-8601 UTC dates, pipe order.

Query: `startDate`, `endDate` (UTC calendar days, inclusive start, exclusive end), `repos`, and a **required**
`granularity` (`daily|weekly|monthly|quarterly|yearly`; 400 when missing, per the batch-1 group decision).

Three concurrent `activities_count` calls (current summary, previous summary, series) through `fetchPipe`
inside `withBucket`. An unknown slug answers 200 with zeros and `[]` after the bucket lookup alone; any
Tinybird fault is 503 `upstream_unavailable`.

### What counts as a code review

The five activity types the Nuxt branch filters on, sent as one `activity_types` array:

| Forge | wire value |
|---|---|
| GitHub | `pull_request-reviewed` |
| GitLab | `merge_request-review-changes-requested`, `merge_request-review-approved` |
| Gerrit | `changeset_comment-created`, `patchset_comment-created` |

The two Gerrit types are also review comments, so this endpoint and `review-comments` (IN-1339) overlap on
them. The OpenAPI operation description lists the types and states the overlap; a test pins both.

## Renames vs the Nuxt handler

| Nuxt (`/api/widget/development/code-review-engagement?metric=code-reviews`) | API |
|---|---|
| `metric=code-reviews` query selector | path leaf `code-reviews` |
| `project=` / `collectionSlug=` | path `{slug}` (collection scope out of scope) |
| `granularity` defaulting to `quarterly` | required, no default |
| `percentageChange` as `Math.abs` | signed |
| `periodFrom`/`periodTo` as `''` when dates omitted | always ISO-8601 UTC |
| bucket `startDate`/`endDate` as `YYYY-MM-DD` | ISO-8601 UTC (`YYYY-MM-DDT00:00:00Z`) |
| bucket `reviews` | kept, typed integer |

## Decisions

- **Parity with the Nuxt params.** The `code-reviews` branch sends `project`, `repos`, `activity_types`,
  `startDate`, `endDate` (and `granularity` on the series) and none of `onlyContributions`,
  `includeCodeContributions`, `includeCollaborations` that `pull-requests` and `issues-resolution` set. The
  pipe's defaults for those flags are outside this repo, so sending them would be a guess that could change
  the count against the widget. A test pins their absence.
- **Resolved current range on the pipe calls.** With both dates omitted the pipe gets `2010-01-01` to today,
  the same period `periodFrom`/`periodTo` report (as `active-days`, `pull-requests`, `commit-activities`).
- **Series rows without both bounds are dropped** (`hasBucketBounds`); a row without an `activityCount`
  reports `0`. One series, so pipe order is kept.

## Tests

`api/tests/development-code-reviews.test.ts`, 37 cases on the shared stubbed-`fetch` pattern: response shape
and key set; the three pipe calls (params, five activity types, absence of the contribution flags, bucket
resolved once, id `0` forwarded, ranges, repos handling, concurrency); signed and null `percentageChange`;
empty results and null bounds; unknown slug; validation (dates, missing and `hourly` granularity, unknown
query key ignored, default range); Tinybird failure table to 503; Cache-Control on 200 and 503; OpenAPI
(tag, every field described, integer counts, query params, activity types and `review-comments` in the
description, absent from `/v1`).

From `api/`: `pnpm test` (455 tests, 21 files), `pnpm tsc-check`, `pnpm lint`, `pnpm format:check` all green.

Two new files, nothing else changed.

Jira: IN-1340


themarolt (approved):
lgtm - checked the five activity types and the pipe params against the nuxt code-reviews branch, they line up, and the shape matches active-days

PR #2248, opened 2026-09-21
test: discover development routes in the shared schema suite

epipav:
## Why

The cross-route suite in `api/tests/common-schemas.test.ts` ran off literal route lists (`seriesRoutes`, `rangeRoutes`, and an integer-summary table). A new endpoint that nobody added to those lists silently skipped the shared `granularity` wording check, the inclusive-start / exclusive-end bounds check, and the unit check. None of the five batch-2 endpoints (#2247, #2245, #2244, #2243, #2246) is in the lists. Raised by @themarolt on #2243. Adding each route in its own PR would put five edits on the same lines of one shared test, the conflict pattern #2241 removed for route registration.

## What

The suite now reads the development routes from the served `/v1-alpha/openapi.json`:
- every development route must take the common range and document the inclusive start and exclusive end
- every route with a `granularity` parameter must mark it required, use the shared enum, and carry `Granularity.description`
- every response property shaped like a period summary (`current`, `previous`, `percentageChange`, `changeValue`), whatever its name, must state a unit on the three values and keep `percentageChange` a nullable number. Integer counts must say `(count ...)`; numbers must say `(seconds)`, `(percent)` or `(percentage points)`. The unit may be followed by a sentence, so nullable notes (merge-lead-time) pass.
- a guard asserts at least five routes, at least one series route and at least one summary were discovered

The module side is already pinned by `tests/v1-alpha-autoload.test.ts` (filename = path leaf), so between the two, a development module cannot be served without being checked.

## Verification

- On `main`: 61 tests in the file, full api suite green, lint, tsc-check, format:check green.
- Against each open batch-2 branch (test file copied in, run, restored): merge-lead-time 67, review-time-by-pr-size 62, code-review-participants 63, review-comments 64, code-reviews 64, all green. So no open PR goes red when this merges, and each gets the checks it was missing.

Jira: IN-1349


api/tests/common-schemas.test.ts
```diff
@@ -406,72 +406,80 @@ describe('periodSummary factory (AC6)', () => {
...
-    spec = res.json<OpenApiDoc>();
-  });
+// Route lists come from the served spec, so a new development module is checked here without
+// anyone adding it to a list; tests/v1-alpha-autoload.test.ts pins the module side.
+const alphaApp = await buildApp();
+await alphaApp.ready();
+const alphaSpec = (
+  await alphaApp.inject({ method: 'GET', url: '/v1-alpha/openapi.json' })
+).json<OpenApiDoc>();
+await alphaApp.close();
```
themarolt: this all runs at collection time now, so anything that throws here takes out all 61 tests in the file rather than failing one - and the `expect(res.statusCode).toBe(200)` that used to guard it is gone, so a non-200 spec shows up as a JSON parse error. same for `paths[...]?.get as OpenApiOperation` on line 426: the cast hides the undefined, so a path under `development/` without a `get` (a POST route later) is a TypeError before any test runs, where the old `toBeDefined()` gave you the route name. could we keep the status assert and filter the discovered paths to the ones that have a `get`?

themarolt (approved):
lgtm - one question on moving the spec fetch to module scope

api/tests/common-schemas.test.ts
epipav: Fixed in 389053a9: the status guard is back as a named error with the body, discovery keeps only paths with a `get` and asserts nothing under the prefix got skipped, and `operation()` throws with the route name instead of casting.

themarolt (approved):
lgtm - the status guard and the $ref resolve cover my earlier point on the module-scope fetch

api/tests/common-schemas.test.ts
```diff
@@ -406,72 +408,128 @@ describe('periodSummary factory (AC6)', () => {
...
+    path.startsWith(developmentPrefix) && item.get
+      ? [[path.slice(developmentPrefix.length), item.get] as const]
+      : [],
+  ),
+);
+const developmentRoutes = [...developmentOperations.keys()].sort();
+const developmentPaths = Object.keys(alphaSpec.paths)
+  .filter((path) => path.startsWith(developmentPrefix))
+  .map((path) => path.slice(developmentPrefix.length))
+  .sort();
```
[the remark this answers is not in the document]
epipav: Fixed in b2620956 the other way round: the comment now states what the check enforces, every development route is a GET, and a path without one fails by name. Building both lists from `get` items would make the assertion always pass and bring the silent skip back.

